PR #21, opened 2019-06-23
[WIP] check_result api change #11

### LINKED ISSUE #11: check_result API
Grader is going through some API changes in #10 .  Thank you all for your input and patience!

Two big motivations are at play: keep it simple (@garrettgman ), keep it clean (@wch ).

---------------------------------------------

After talking it through with @blairj09, I think we have a clean, simple solution for `check_result`.
* Two check object functions: `passes` and `fails`.  (These names do not class with `checkmate` or `testthat`.)
* `passes` and `fails` can take:
  * a Value - equivalent to `~ identical(., value)`
  * a Formula - Evaluate the formula with `.` representing the user result
    * By using a formula (purrrr style) with `.`, there is no non-standard evaluation trickery to understand (such as a random `ans` being available).  As long as we state that `.` is the user result, the validation can be processed however the user writes.
    * Should return `TRUE` / `FALSE`
    * Throwing will cause an error in computing (bad for author)
  * a Function that takes an user result argument
    * Should return `TRUE` / `FALSE`
    * Throwing will cause an error in computing (bad for author)
* Once an equality (against a value) is found or `TRUE` (from a formula / function) is returned, computation will stop.  The respective message will be displayed as a correct (`passes`) or incorrect (`fails`) result.
* `correct` or `incorrect` params will be the default message for any `passes` or `fails` that do not provide a message.  They will be `glue`ed with appropriate information.

Example:

```r
check_result(
  fails(5, "bad answer!"),
  fails(~ . == 5, "bad answer!"),
  fails(~ isTRUE(all.equal(., 5)), "bad answer!"),
  fails(~ . %in% c(5,7), "bad answer!"),
  fails(function(x) x == 5, "bad answer!"),
  fails(function(x) x == 5),
  fails(somepkg::some_method, "bad answer!"), # function(x) ...

  passes(4, "good answer!"),
  passes(~ . == 4, "good answer!"),
  passes(~ . %in% c(4,6), "good answer!"),
  passes(function(x) x == 4, "good answer!"),
  passes(function(x) x == 4),
  passes(somepkg::some_method, "good answer!"), # function(x) ...

  correct = "You got it right!",
  incorrect = "You got it wrong."
)
```

---------------------------------------------

`check_result` will be kept separate from `test_result`.  All tests will be executed for `test_result`. 
* `test_result` will take
  *  a `test` object which will only accept formulas or functions as defined above (no values).  The message supplied will be provided if any throw occurs.
  * a Formula - the thrown error will be reported
  * a Function - the thrown error will be reported
* `correct` and `incorrect` will be `glue`ed with appropriate information.

Example:
```r
test_result(

  ~ checkmate::assert_function(., args = "x"),
  test(~ checkmate::assert_function(., args = "x"), "Make sure to have an 'x' argument!"),
  somepkg::some_expectation, # function(x) ...
  pryr::partial(checkmate::assert_function, args = "x"),

  correct = "{num_correct} / {num_total} correct! {random_praise()}",
  incorrect = "{num_correct} / {num_total} correct. {random_encourage()}"
)
```

---------------------------------------------

Again, these are not set in stone... just a followup from discussions over the past 48 hrs.

cc @gvwilson

chendaniely:
When this PR gets merged it Fixes #11.

@schloerke @garrettgman I believe I have all the working parts in there, but I can't seem to get the actual `pass_if` API working the way it was described in #11.

In the `pass_if` function signature, should the `user_answer, grader_args = list(), learnr_args = list()` all be passed as `...`?


chendaniely:
@schloerke I have a question of getting `fail_if` to work/pass tests:

The following test will fail
```r
  expect_error(
    check_result(
        fail_if(~ .result == 5, "You were supposed to do this other thing!"),
        grader_args = list(solution_quo = quote(5)),
        learnr_args = list(envir_prep = new.env())
    )
  )
```

because I need one correct result to be passed (from `check_result`)

```r
  if (!any(vapply(results, `[[`, logical(1), "correct"))) {
    stop("At least one correct result must be provided")
  }
```

The problem is, that the match exists. So should I be taking out that `if` statement? I'm not sure why it's there? Was it from the previous `result` api where one of the result values needed to be 'correct'?
The test will pass if I comment out the lines in `check_result`

R/check_result.R
```diff
@@ -30,24 +31,39 @@ check_result <- function(
...
 
   # init final answer as not found
   final_result <- graded(correct = FALSE, NULL)
   found_match <- FALSE
   for (resu in results) {
-    if (identical(resu$x, user_answer)) {
-      final_result <- resu
-      found_match <- TRUE
-      break
+    # don't break the result api... yet
```
schloerke: break it

R/check_result.R
```diff
@@ -16,6 +16,7 @@
 #' @template learnr_args
 #' @param ... ignored
 #'
+#' TODO eventuall will return a grader_condition object from condi
```
schloerke: returning a `grader_graded` is correct.  While it may contain similar info, the mental model is different and should be kept separate.

* A `condi` is a condition used to determine a `graded` value

R/check_result.R
```diff
@@ -30,24 +31,39 @@ check_result <- function(
   learnr_args = list() # provided by `grade_learnr`
 ) {
   results <- list(...)
-  chkm8_item_class(results, "grader_result")
+  #chkm8_item_class(results, "grader_result") ## TODO: make this "grader_condition" # nolint
```
schloerke: +1

R/pass_fail_condi.R
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#' Evaluates a condition
+#' @returns Boolean of the condition
+#' @export
+evaluate_condi <- function(condi, grader_args, learnr_args) {
+  checkmate::assert_class(condi, "grader_condition")
+  switch(condi$type,
+         "formula" = evaluate_condi_formula(condi$x, grader_args$solution_quo, learnr_args$envir_prep), # nolint
+         "function" = evaluate_condi_function(condi$x, grader_args$solution_quo),
```
schloerke: might need envir_prep. idk for sure. Not the priority right now

R/check_result.R
```diff
@@ -30,24 +31,39 @@ check_result <- function(
...
+      }
+    } else if (class(resu) == "grader_condition") {
+      # if formula then eval with more infomration, needs to return TRUE
+      # if x is is a funciton execute with last user_answer (fxn needs to return TRUE/FALSE,
+      # e.g., an assert statement)
+      # else do this below
+      evaluated_condi <- evaluate_condi(resu, grader_args, learnr_args)
+      if (evaluated_condi) {
+        found_match <- TRUE
+        final_result <- list(correct = resu$correct,
```
schloerke: maybe generate a `graded` value from the `resu`?

schloerke: Similar to `final_result <- graded(correct = FALSE, NULL)`

R/pass_fail_condi.R
```diff
@@ -0,0 +1,65 @@
...
+  class(ret) <- "grader_condition"
+  ret
+}
+
+#' Evaluates a condition
+#' @returns Boolean of the condition
+#' @export
+evaluate_condi <- function(condi, grader_args, learnr_args) {
+  checkmate::assert_class(condi, "grader_condition")
+  switch(condi$type,
```
schloerke: wrap this in a `try(..., silent = TRUE)`. If the result inherits a `try-error`, then return a `graded(FALSE, NULL)`.  

(Comparisons could cause an error. Function evaluation could cause an error. Good to be safe.)

R/pass_fail_condi.R
```diff
@@ -0,0 +1,65 @@
...
+    message = message,
+    correct = correct,
+    type = type
+  )
+  class(ret) <- "grader_condition"
+  ret
+}
+
+#' Evaluates a condition
+#' @returns Boolean of the condition
```
schloerke: should return `graded` value

R/pass_fail_condi.R
schloerke: inside the `try`, If the value is TRUE, return  `graded(condi$correct, condi$message)`.  Otherwise return NULL

R/pass_fail_condi.R
schloerke: ... or `NULL` if the condition is false

R/pass_fail_condi.R
schloerke: If the value is not a boolean or `graded` output, throw an error.

chendaniely: The `NULL` is causing a few problems... Correct me if I'm wrong and might be thinking about this the wrong way...

Essentially `res` acts as a "did I find a match?" and if there was an error in running the `x` we return `graded(correct = FALSE, message = NULL)`.

So, we shouldn't need to check if `res` is `NULL` since that doesn't seem will ever happen (it should only be a T/F value or a `graded` object).

Next, since `res` is a logical:
if it's true (we found a match), then we return `return(graded(correct = condi$correct, message = condi$message))`
which is fine... but if we do not find a match we should return something else (maybe `NULL` makes sense only in this location?).

I say this because when we leave `evaluate_condi` and go back to `check_result`, `evaluated_condi` needs to account for a matched `fail_if` call. Right now if `fail_if` does match, we never do anything about it

```r
if (!is.null(evaluate_condi) && evaluated_condi$correct) {
          found_match <- TRUE
          final_result <- evaluated_condi
          break
        }
```

R/pass_fail_condi.R
chendaniely: not sure if this is needed since the function is called within check_result. but can come back to it if the instructor puts in a function that makes calls to variables in the environment.

R/pass_fail_condi.R
chendaniely: discussing no longer relevant. we removed the tryCatch and if the student's solution is wrong, we return `NULL` this lets us check for a found match without having to also check for a correct solution

chendaniely:
only question I had was wrt 88cbc94. 
I had to make that change to get the passing message through, but that means in `check_result.R`

```r
 message <- glue::glue_data(
    list(
      matched = found_match,
      correct = final_result$correct,
      message = final_result$message
    ),
    {if (final_result$correct) correct else incorrect} # nolint
  )
```

does not get used...

R/check_result.R
```diff
@@ -62,6 +64,6 @@ check_result <- function(
 
   return(graded(
     correct = final_result$correct,
-    message = message
+    message = final_result$message
```
schloerke: Why not keep it as `message`?

This value is created from the `final_result$message` in _new line 60_.   By using the glue statement, encouragement and praise can be added in addition to the passing / failure message provided by `final_result`.

schloerke: @chendaniely

tests/testthat/test_check_result_condi.R
```diff
@@ -0,0 +1,121 @@
...
+    check_result(
+      fail_if(~ .result == 5, "You were supposed to do this other thing!"),
+      fail_if(~ .result == 10, "You were supposed to do this other thing!"),
+      grader_args = list(),
+      learnr_args = list(last_value = quote(5), envir_prep = new.env())
+    )
+  )
+})
+
+test_that("Spots differences in atomics -- formuula", {
```
schloerke: `formula`

R/check_result.R
```diff
@@ -30,23 +31,24 @@ check_result <- function(
...
 ) {
   results <- list(...)
-  chkm8_item_class(results, "grader_result")
+  chkm8_item_class(results, "grader_condition")
   chkm8_single_character(correct)
   chkm8_single_character(incorrect)
 
   if (!any(vapply(results, `[[`, logical(1), "correct"))) {
-    stop("At least one correct result must be provided")
+   stop("At least one correct result must be provided")
```
schloerke: add the space back

R/graded.R
```diff
@@ -0,0 +1,19 @@
...
+#'\code{*-check} chunk when used with \code{link{grade_learnr}}.
+#'
+#' @param message A character string of the message to be displayed.
+#' @param correct A boolean value of whether or not the checked code is correct.
+#' @export
+graded <- function(correct, message = NULL) {
+  chkm8_single_character(message)
+  checkmate::expect_logical(correct, any.missing = FALSE, len = 1, null.ok = FALSE)
+
+  structure(class = "grader_graded",
```
schloerke: Can we do as below?

```r
ret <- list(
  message = message %||% "",
  correct = correct
)
class(ret) <- "grader_graded"
ret
```

tests/testthat/test_check_result_condi.R
```diff
@@ -0,0 +1,121 @@
...
+  expect_s3_class(x, "grader_graded")
+  expect_equal(x$correct, correct)
+  expect_true(grepl(message, paste0(x$message, collapse = ""), fixed = TRUE))
+}
+
+test_that("Provide a passing solution. Give the students a fighting chance!", {
+  testthat::expect_error(
+    check_result(
+      grader_args = list(),
+      learnr_args = list(last_value = quote(5), envir_prep = new.env())
```
schloerke: I believe all of the values for `last_value` should be _raw_ values.  Such as `5` (not `quote(5)`)

schloerke: (I know it doesn't matter for these tests in particular, but good to be consistent.)

tests/testthat/test_condi.R
```diff
@@ -0,0 +1,71 @@
...
+    expect_equal(x$message, message)
+    expect_false(x$correct)
+}
+
+grader_args <- list()
+learnr_args <- list(last_value = quote(5), envir_prep = new.env())
+condi_formula_t <- condition(~ .result == 5, message = "my correct message", correct = TRUE)
+condi_formula_f <- condition(~ .result == 1, message = "my error message", correct = FALSE)
+
+context("Check condi")
```
schloerke: this should be at the top of the setup (so if an error occurs we can see what context it is in)

tests/testthat/test_grade_learnr.R
```diff
@@ -44,7 +44,20 @@ test_grade_learnr <- function(user_code,
                last_value = last_value)
 }
 
-test_that("grade learnr", {
+test_that("Grade learnr check_result", {
+  expect_correct(
+    test_grade_learnr(
+      user_code = "4",
+      check_code = "check_result(
+          pass_if(~ .result == 4, 'yes. you did it')
```
chendaniely: @schloerke this is why I changed that `graded` value from `message` to `final_result$message` I was getting a random praise otherwise.

However, i need to change it back to `message` becuase the message does not show up when I actually test it out in a learnr tutorial

R/check_result.R
chendaniely: see comment: https://github.com/rstudio-education/grader/pull/21/files#r300204016

tests/testthat/test_grade_learnr.R
schloerke: Can test with something like

```r
check_result(correct = "{message}", pass_if(~TRUE))
```

chendaniely:
If this review is good, can we merge this and work on #29 separately?

R/check_result.R
```diff
@@ -30,23 +31,24 @@ check_result <- function(
...
   final_result <- graded(correct = FALSE, NULL)
   found_match <- FALSE
+
   for (resu in results) {
-    if (identical(resu$x, user_answer)) {
-      final_result <- resu
+    evaluated_condi <- evaluate_condi(resu, grader_args, learnr_args)
+    if (! is.null(evaluated_condi)) {
       found_match <- TRUE
+      final_result <- list(correct = resu$correct,
```
schloerke: If the`evaluated_condi` is not `NULL`, it should be a `graded` result.

Use 

```r
graded(correct = evaluated_condi$correct, message = evaluated_condi$message)
```

R/pass_fail_condi.R
```diff
@@ -0,0 +1,100 @@
...
+#' Evaluates a condition
+#'
+#' @param condi a \code{grader} \code{\link{condition}} object
+#' @param grader_args at minimum, a list that just contains the value for \code{solution_quo}
+#' @param learnr_args at minimum, a list that just contains the value for \code{envir_prep}
+#'
+#' @return a \code{graded} value if \code{condi$x} is \code{TRUE} or
+#'   \code{NULL} if \code{condi$x} is \code{FALSE}
+#' @export
+evaluate_condi <- function(condi, grader_args, learnr_args) {
```
schloerke: change to `evaluate_condition` to match `condition`

R/check_result.R
chendaniely: fixes #29

tests/testthat/test_grade_learnr.R
```diff
@@ -50,10 +50,10 @@ test_that("Grade learnr check_result", {
       user_code = "4",
       check_code = "check_result(
           pass_if(~ .result == 4, 'yes. you did it'),
-          correct = 'this other correct message'
+          correct = 'this other correct message. {message}'
```
chendaniely: oh! that's how that worked : |

chendaniely:
This PR also Fixes #29

PR #10, opened 2019-04-22
Add checking methods: `check_code`, `check_result`, and `test_result`

schloerke:
Fixes #7 
Fixes #8 
Fixes #9
Fixes https://github.com/rstudio/learnr/issues/204

- Exercise check methods
  - [x] check_code (formerly check_strict)
  - [x] check_result
  - [x] test_result
- [x] pkgdown support
- [x] travis support
  - [x] auto-deploy pkgdown from master
- [x] codecov support
- [x] Update readme

## Testing notes

* Visit 
  * https://rstudio-education.github.io/grader/
  * https://github.com/rstudio-education/grader/tree/grade_exercise_answer
*  Run grading tutorial. 
* Submit correct and incorrect solutions for all three methods.

```r
remotes::install_github("rstudio-education/grader#10")
learnr::run_tutorial("grading-demo", "grader")
```

PR #22, opened 2019-06-27
[wontfix] Resurrect strict_check

chendaniely:
`strict_check` got removed when #10 was merged. This PR tries to reset everything to make sure all the grading features are in the library.


chendaniely:
@garrettgman there was a bit of confusion, the `strict_check` function you wrote was moved and renamed to `check_code`. This way we have a `check_results` and a `check_code` for the various ways to 'grade' code.

I'll close this PR after our meeting so we're all on the same page.

chendaniely:
no need for this PR since `strict_check` was renamed and modified into `check_code`

PR #40, opened 2019-07-17
Check result messages

### LINKED ISSUE #37: Allow pass_if and fail_if to pass in a NULL message
be able to use

`grader::pass_if(~ .result == correct_df)` instead of only

`grader::pass_if(~ .result == correct_df, message = "1337")`

chendaniely:
Fixes #37


chendaniely:
Travis is failing on the `oldrel` version of `R`. Is this needed?

chendaniely:
@schloerke @garrettgman I'm thinking of taking the unit tests out of this PR because allowing instructors to not pass in a message into `pass_if` and `fail_if` is a good chance.

We need to do some internal refactoring on how the messages are created anyway, so the unittests wouldn't be relevant to the `message = NULL` default anyway.

PR #62, opened 2019-08-13
Doc fixes

chendaniely:
Documentation updates and fixes

Does not fix #60 that's in another PR (#63)

PR #63, opened 2019-08-14
Test result api fixes#60

chendaniely:
This will be rebased after #62 gets merged.

Implements the changes for `test_result`


chendaniely:
This will be rebased after #62 gets merged

chendaniely:
`correct` and `incorrect` should be used to return when all tests pass or not, respectively

R/glue.R
```diff
@@ -19,12 +19,21 @@ glue_message <- function(
...
 
   ret <- glue::glue_data(params, glue_expression)
   return(ret)
 }
+
+clean_char_names <- function(char_name, params) {
+  x <- params[[char_name]] %||% "" # convert NULL strings to "" to work with glue
+
+  tryCatch({
+    chkm8_single_character(x, char_name)
```
schloerke: this tryCatch setup seems like a fix for a different problem.

Maybe run `x <- as.character(x)` after `x` is defined on line 29? Then call `chkm8_single_character(x, char_name)`

R/test_result.R
```diff
@@ -1,127 +1,143 @@
...
+#' @template incorrect
 #' @template grader_args
 #' @template learnr_args
-#' @param ... ignored
+#' @param glue_correct A glue string that returns the final correct message displayed.
+#'    Defaults to \code{getOption("gradethis_glue_correct_test")}, e.g.,
+#'    2/2 correct! Absolutely fabulous!.
+#' @param glue_incorrect A glue string that returns the final correct message displayed.
+#'    Defaults to \code{getOption("gradethis_glue_correct_test")}, e.g.,
+#'    1/2 correct! Try it again; next time's the charm!.
```
schloerke: missing code and quotes.

`\code{"1/2 correct! ... "}`

R/test_result.R
```diff
@@ -1,127 +1,143 @@
...
-  tests <- grader_tests(...)
-  chkm8_class(tests, "grader_tests")
-  chkm8_single_character(correct)
-  chkm8_single_character(incorrect)
 
-  user_answer <- get_user_code(grader_args$user_quo)
+  conditions <- list(...)
+  chkm8_item_class(conditions, "grader_condition")
+
+  test_results <- purrr::map(conditions, pass_fail_condition_modify,
```
schloerke: Is this the only place we use `purrr`? While I like the tidyverse coding style, it is an unnecessary dependency.  (Use `lapply`)

R/test_result.R
```diff
@@ -1,127 +1,143 @@
...
+  # so we would need to flip the graded class correct status
+  if (condi$correct) { # evaluating a pass_if condition # nolint
+    # if a pass_if returns a NULL, it means the condition evaluated FALSE, which is a bad thing
+    # if it is NULL, we give it an "incorrect" graded class value
+    # if it is "correct", we keep the "correct" graded class value
+    evaluated_condi <- evaluated_condi %||% graded(correct = FALSE, message = NULL)
+    return(evaluated_condi)
+
+  } else { # evaluating a fail_if condition # nolint
+    if (is.null(evaluated_condi)) {
```
schloerke: I believe this if/else can be replaced with 

```r
evaluated_condi <- evaluated_condi %||% graded(correct = TRUE, message = NULL)
```

A _passing_ `fail_if` returns a `graded(correct = FALSE, message = MESSAGE)` object.

R/glue.R
chendaniely: it came up because of `test_result`'s num_correct / total_tests calculation since those values are numbers and threw an error. I initially thought of not using `tryCatch`, but I figure this would be a placeholder for some other edge case in the future?

R/test_result.R
chendaniely: @garrettgman uses it in his `unpipe` and `order_calls` files. We could re-write those, the main one that isn't a straight `lapply` is `purrr::safely`

R/glue.R
schloerke: Maybe we add in a new thing of starting with `.num_` is numeric? (Like how we did for bools) 

Otherwise remove the `tryCatch` and just use the `error` portion of it.

R/test_result.R
schloerke: ok.  keep the `purrr`. Thanks

R/test_result.R
chendaniely: oh you're right! I just wrote it super explicitly to not make my brain hurt.
I should update the docs about that too...

R/evaluate_condition.R
```diff
@@ -1,4 +1,4 @@
-#' Evaluates a condition
+#' Evaluates a conditio
```
schloerke: missing `n`

archive/view_tutorial.R
```diff
@@ -12,7 +12,10 @@
 #'
 #' TODO need to export when feature is implement in RStudio IDE
 #' @noRd
+<<<<<<< HEAD:archive/view_tutorial.R
```
schloerke: remove

R/glue.R
chendaniely: Just realized, `chkm8_single_character` may not be the correct function to use here.
We just want a valid value that can be used in `glue::glue` but glue can take numeric values, so we should'nt just be checking for characters.

we really just want each element in params to be any atmoic of length 1, and not something like a list, right?

R/test_result.R
chendaniely: Will make the change in #68

R/glue.R
chendaniely: moving to use `expect_atomic`: https://mllg.github.io/checkmate/reference/checkAtomic.html

PR #74, opened 2019-08-22
Add function formals when comparing and standarising code for comparison

chendaniely:


chendaniely:
@schloerke seems like the unittest needed me to manually pass in the `env`, just want to make sure that looks right before I integrate into gradethis

schloerke:
@chendaniely 
Passing the env is normal for functions like these.  You have a good default as I don't think we can extract the `env` from the `code` argument.

chendaniely:
build seems to be failing becuase of a `R CMD check`

I get this now when I run `devtools::check()` on my computer, didn't seem to happen before. nor do I understand why the `tidyr::` calls in billboard.R or in the tests (they're quoted expressions) would cause this...


```
-- R CMD check results ------------------------------- gradethis 0.1.0.9002 ----
Duration: 24.2s

> checking for unstated dependencies in 'tests' ... WARNING
  '::' or ':::' import not declared from: 'tidyr'

0 errors v | 1 warning x | 0 notes v
Error: R CMD check found WARNINGs
Execution halted

Exited with status 1.
```

tests/testthat/test_check_code.R
```diff
@@ -100,9 +104,11 @@ test_that("Mentions only first non-matching element", {
 })
 
 test_that("Spots differences in argument names", {
-  a <- quote(mean(1:10, trim = 1, na.rm = TRUE))
-  b <- quote(mean(1:10, 1, TRUE))
-  c <- quote(mean(1:10, cut = 1, na.rm = TRUE))
+  test_fn <<- function(x, y = 1, z = 2, ...) {return(1)}
```
chendaniely: @schloerke example of global `<<-` assignment operator to fix tests

PR #85, opened 2019-12-16
Adds tidyr as a suggests, disables lintr

garrettgman:


chendaniely:
`tidyr` is just used for one of the tests.
https://github.com/rstudio-education/gradethis/blob/6a6b21c2a87513bc3453e478a62f3cbd426a3d1a/tests/testthat/test_check_code.R#L213
It's not actually used in the code for the package.

the `billboard.R` script (https://github.com/rstudio-education/gradethis/blob/master/scripts/billboard.R) is used to create the dataframe structure used in the test (https://github.com/rstudio-education/gradethis/blob/master/tests/testthat/billboard.R)

Having said that.. no idea why it's being requested in `CMD CHECK` since it's being called in `quote` and not evaluated

schloerke:
@chendaniely I believe it's being found with static code check.  Ex: `grep("::", code, fixed = TRUE)`

But also, the "standardize formals" function still needs to find the function being used.  So the code (pkg) needs to exist.

chendaniely:
Oh that would make sense why it's asking for it all of a sudden...
Should we change the test to something that's an expression that's longer than `60L`?

I only remember going through that entire `tidyr` billboard script to reduce the dependencies used by the package.

PR #92, opened 2020-03-13
Plus handler

garrettgman:
Eric, Jeff, and I noticed that `detect_mistakes` behaved weirdly in the following way: when comparing a user submitted call with a solution's call, `detect_mistakes` first counted how many arguments appeared in each call  and alerted the user if there was a surplus argument. This resulted in irrelevant/confusing messages if the user and the solution used two different calls (e.g. as described in #84, `sqrt` vs `+`). I tweaked `detect_mistakes` to fix this behavior.


R/message_generators.R
```diff
@@ -3,59 +3,237 @@
...
+  # "In g(1, h(i(1))), Your call to h() should include b",
+  # "as one of its arguments. You may have referred to it ",
+  # "in a different way, or left out an important argument name."
+  
+  # "{intro}Your call to {this_call} should include {that_name} ",
+  # "as one of its arguments. You may have referred to it ",
+  # "in a different way, or left out an important argument name."
+  
+  intro <- build_intro(.call = enclosing_call, .arg = enclosing_arg)
+  intro <- ifelse(intro == "", "Y", paste0(intro, "y"))
```
schloerke: Store this as a `your_char` variable and add it to the glue statement.  `"{intro}{your_char}our..."`

R/message_generators.R
```diff
@@ -64,3 +242,16 @@ prep <- function(text) {
   if (!is.character(text)) text <- deparse_to_string(text)
   text
 }
+
+build_intro <- function(.call = NULL, .arg = NULL) {
+  
+  if(!is.null(.call)) {
+    .call <- deparse_to_string(.call)
+    if (!is.null(.arg) && .arg != "")
```
schloerke: Add `{``}` here.

schloerke: Change `.arg != ""` to be `!identical(.arg, "")`

tests/testthat/test_sanity_checks.R
```diff
@@ -0,0 +1,566 @@
+context("Check sanity messages")
+
+a <<- function(x) x
+b <<- function(x) x
+f <<- function(x, y) x + y
+g <<- function(x, y = a(1)) x + y
```
schloerke: Only need single arrow assignments. `a <- function....`

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+  solution_args <- as.list(solution)
+  real_solution_names <- solution_names[solution_names != ""]
+  
+  ## Remove exact matches from further scrutiny
+  for (name in real_user_names) {
+    if (name %in% real_solution_names) {
+      user_args[[name]] <- NULL
+      solution_args[[name]] <- NULL
+    }
+  }
```
schloerke: Make sure this passes with `f(a = 1, a = 2)` where `f(a = 1)` is correct

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+  
+  ## Check remaining arguments for partial matches
+  user_names <- rlang::names2(user_args)
+  remaining_user_names <- user_names[user_names != ""]
```
schloerke: Let's make a helper routine for this

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+  
+  if (length(remaining_user_names) > 0) {
+    solution_names <- rlang::names2(solution_args)
+    remaining_solution_names <- solution_names[solution_names != ""]
+    
+    ## Do any solution names partially match multiple user names?
+    pmatches_per_formal <- function(solution_name) {
+      sum(startsWith(solution_name, remaining_user_names))
+    }
```
schloerke: Check out `match.arg`

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+  
+  ## Check remaining arguments for partial matches
+  user_names <- rlang::names2(user_args)
+  remaining_user_names <- user_names[user_names != ""]
+  
+  if (length(remaining_user_names) > 0) {
+    solution_names <- rlang::names2(solution_args)
+    remaining_solution_names <- solution_names[solution_names != ""]
+    
+    ## Do any solution names partially match multiple user names?
```
schloerke: `Do any non-matched solution names...`

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+    }
+  }
+  
+  ## Check remaining arguments for partial matches
+  user_names <- rlang::names2(user_args)
+  remaining_user_names <- user_names[user_names != ""]
+  
+  if (length(remaining_user_names) > 0) {
+    solution_names <- rlang::names2(solution_args)
+    remaining_solution_names <- solution_names[solution_names != ""]
```
schloerke: Move these two lines above the if statement

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
-           )
-         )
-      }
-      # if user[i] solution[i] not same: isolate_mismatch_2
-      if (!identical(user[[i]], solution[[i]])) {
-        return(
-          detect_mistakes(user[[i]], solution[[i]], env = env)
+  # Check for unused unnamed arguments
+  # Remove used names
+  for (name in remaining_user_names) {
```
garrettgman: This fails to remove names that are successfully partially matched. (It is redoing the removal of EXACT matches.

garrettgman: Also this should go inside the if bracket at ~ line 138

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
-             this = user[[i]],
-             this_name = user_names[[i]]
-           )
-         )
-      }
-      # if user[i] solution[i] not same: isolate_mismatch_2
-      if (!identical(user[[i]], solution[[i]])) {
-        return(
-          detect_mistakes(user[[i]], solution[[i]], env = env)
+  # Check for unused unnamed arguments
```
schloerke: Add comment about how all user args will be unnamed arguments from this point on

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+        enclosing_arg = enclosing_arg
+      )
+    )
+  }
+  
+  # 7. Check that every named argument in the solution matches every
+  #    correspondingly named argument in the user code. We know each 
+  #    has a match because of Step 5.
+  user_args <- as.list(user[-1])         # remove the call
+  solution_args <- as.list(solution[-1]) # remove the call
```
schloerke: Move the `[1]` outside `as.list()`

R/detect_mistakes.R
```diff
@@ -1,281 +1,302 @@
...
+          enclosing_call = enclosing_call,
+          enclosing_arg = enclosing_arg
+        )
+      )
+      
+    # The user argument has a matching solution argument, are they identical?
+    } else if (!identical(user_args[[i]], solution_args[[i]])) {
+      name <- rlang::names2(user_args[i])
+      if (!(name %in% submitted_names)) name <- ""
+      return(
```
schloerke: Only return if the result is `!is.null(res)`.  This applies to Section 7 above. ~line 217

schloerke: Have the first pair of unnamed args match. Have the second pair of unstandardized args mismatch.  Currently `detect_mistakes` will say _valid_. It should be _invalid_.

tests/testthat/test_sanity_checks.R
garrettgman: Do I? Because changing from `<<-` to `<-` causes the tests to fail. After probing, the cause of the failure is testthat's inability to find the function definitions (and the formals and default values associated with them) when running the tests with `<-`. I'm not sure how testthat scoping works (which is a terrible scenario for using `<<-` !), but `<<-` does fix things.

garrettgman: I assume the fix is taking advantage of `<<-`'s behavior to assign things in the global environment environment if they are not found on the search path. This doesn't seem to overwrite what is loaded in my global environment, so I abduce that testthat runs tests in a new clean R session.

R/detect_mistakes.R
garrettgman: Fixed in https://github.com/rstudio-education/gradethis/pull/92/commits/6cc9595c931505cee0f5c9618c58530a559a60b8

R/detect_mistakes.R
garrettgman: `match.arg` won't work here because it throws an error if the solution name is not in the `remaining_user_names`

R/detect_mistakes.R
garrettgman: Fixed in https://github.com/rstudio-education/gradethis/pull/92/commits/e4e8efc50e5c316314a05f881e09313c36dffee9

tests/testthat/test_sanity_checks.R
schloerke: It still seems very sketchy and we should be able to get around this.  I'm ok leaving it for now. We can add a separate issue for it.  (Reported in https://github.com/rstudio-education/gradethis/issues/104)

PR #115, opened 2020-05-27
Introduce `find_matching_condition()`

trestletech:
`find_matching_condition()` has been extracted from `grade_result()`. This allows us to determine whether or not a result actually matches any of the rules, or if we're just defaulting to a failure.

**Notes**:

 - Best to review commit-by-commit. The second commit modernizes a bunch of Rd files to bring them up to a more recent roxygen version.


R/grade_result.R
```diff
@@ -47,18 +77,6 @@ grade_result <- function(
...
-    .correct = correct,
-    .incorrect = incorrect
-  )
-
-  return(graded(
-    correct = final_result$correct,
-    message = message
-  ))
+  
+  list(matching_conditional = found_match, result = final_result)
```
trestletech, line 81: find_first_condition

return instead of a list, just final_result if found, or NULL otherwise.

Other functions update to handle that signature.

R/grade_result.R
```diff
@@ -27,30 +27,17 @@ grade_result <- function(
...
-      final_result <- evaluated_condi
-      found_match <- TRUE
-      break
-    }
+  args <- list(..., grader_args = grader_args, learnr_args = learnr_args)
+  mat <- do.call(find_first_condition, args)
+  
+  final_result <- mat
+  if (is.null(mat)){
+    final_result <- list(correct = FALSE, matching_conditional = FALSE, message="")
```
trestletech, line 35: I'm guessing here. I'm not sure what the default value should be.

trestletech:
@schloerke I'm finally getting back to this; sorry for the delay.  Just pushed a commit that addresses what we discussed.

But I'm realizing that this is pretty close to the work that we'd discussed around #118 so if you'd rather just incorporate this branch into your work there and merge it all at once, that's fine by me.

Just keep me posted 'cause I do have a client that's calling these functions so I'll need to make sure that they're still exported in some form and that I know what the behavior's going to be. Thanks!

R/grade_result.R
```diff
@@ -62,3 +49,30 @@ grade_result <- function(
...
+#' `NULL`.
+#' @export
+find_first_condition <- function(..., grader_args, learnr_args){
+  
+  results <- list(...)
+  chkm8_item_class(results, "grader_condition")
+  
+  if (!any(vapply(results, `[[`, logical(1), "correct"))) {
+    stop("At least one correct result must be provided")
+  }
```
cpsievert, line 68: @trestletech if you don't mind, maybe we should take on https://github.com/rstudio-education/gradethis/issues/118 as a part of this PR, which *I think* would mean the following change:

```suggestion
  if (!any(vapply(results, `[[`, logical(1), "correct"))) {
    results <- c(results, list(pass_if(TRUE)))
  }
```

cpsievert:
@trestletech what if we just attached a special class to this default final result?

https://github.com/rstudio-education/gradethis/blob/c19a00ae043138ef18e4494e3b30616b8356cab5/R/grade_result.R#L39

That'd be a much simpler change and wouldn't require exporting an additional function

cpsievert:
Will be handled in #125

PR #124, opened 2020-07-22
Don't require pass_if() in grade_results()

### LINKED ISSUE #118: grade_result() should not require a pass_if()
`grade_result()` returns an error message if it is called without at least one `pass_if()` statement. Here's an example from the `grading_demo()`, notice that I commented out the `pass_if()` statement to create this error

Screen Shot 2020-06-12 at 9 54 02 AM

We should allow the user to call `grade_result()` without a `pass_if()` statement to simplify the case where there are many correct answers and the grader wants to screen for specific wrong answers. You can do this now by including `pass_if(~FALSE)` in the `grade_result()` call as a hack:

Screen Shot 2020-06-12 at 9 57 46 AM

p.s. You can accomplish this by switching to `grade_conditions()` and flipping the logical sign of your tests, but that:
1. Seems like unnecessary mental gymnastics
2. Complicates my efforts to write a function that uses `grade_results()` to handle both cases 😈

cpsievert:
Closes #115 
Closes #118

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -27,7 +27,7 @@ Description: Pairing with the 'learnr' R package,
 License: GPL-3
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 6.1.1
+RoxygenNote: 7.1.1
 Imports:
     purrr,
     rlang,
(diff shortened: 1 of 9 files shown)


garrettgman:
@cpsievert the code changes make sense to me. Do you foresee any big downstream effects?

PR #126, opened 2020-07-23
evaluate_condi_formula() should evaluate using envir_result

### LINKED ISSUE #101: `eval_condi_function()` should execute function in student environment
`eval_condi_function()` should execute function in student environment. The below should evaluate in the known `env` with `eval_tidy()`:

https://github.com/rstudio-education/gradethis/blob/6671ceacae7b779116128f424f436a279cca44b6/R/evaluate_condition.R#L70

### LINKED ISSUE #121: Evaluate conditions in `envir_result` rather than `envir_prep`
Currently, when evaluating formula via `evaluate_condi_formula`, the evaluating environment is `envir_prep` which is the environment before any student code is run. This means the only way to access any output from the user is from `.result` and that only gives the last line. If you pass `envir_result` to `evaluate_condi_formula()` instead of `envir_prep` then the conditions can access evaluated environment of the user. This allows for better checks like `pass_if(~ exists(obj))` where `obj` is something that the student is supposed to create. I can make a PR, if it's helpful.

cpsievert:
Fixes #101
Fixes #121 

Before this change entering `a<-1` would result in a meaningless error. After this change, entering `a<-1` will pass the test

````
---
title: "Checking errors"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)	
library(gradethis)
tutorial_options(
  exercise.checker = grade_learnr
)
```

* Please enter `a <- 1`

```{r grade_error, exercise = TRUE}

```


```{r grade_error-check}
grade_result(pass_if(~a==1))
```
````

### DIFF
--- a/R/evaluate_condition.R
+++ b/R/evaluate_condition.R
@@ -23,7 +23,7 @@ evaluate_condition <- function(condition, grader_args, learnr_args) {
   err_msg <- NULL
   res <- tryCatch({
     switch(condition$type,
-           "formula" = evaluate_condi_formula(condition$x, learnr_args$last_value, learnr_args$envir_prep), # nolint
+           "formula" = evaluate_condi_formula(condition$x, learnr_args$last_value, learnr_args$envir_result), # nolint
            "function" = evaluate_condi_function(condition$x, learnr_args$last_value),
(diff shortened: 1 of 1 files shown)


cpsievert:
TODO: why are there test failures?

PR #146, opened 2020-09-03
Fix grade_code message for <- call

nischalshrestha:
# Description

This PR fixes issue https://github.com/rstudio-education/gradethis/issues/137 where the feedback for not using an assign call produces a confusing message. 

If user code is `123` and solution code is `a <- sample(1:6, size = 1)`, we would get:

`"I expected NULL <- NULL where you wrote 123."`

# Fix

Handle extracting the `<-` out of solution code in `message_generators.R` by using `[[` to extract code and adding an extra pattern grep for `<-` as well. This works but feels a bit hacky and starts to touch on how we should grade expressions with or without infix operators.

---

Now, if user code is `123` 

And solution code is `a <- sample(1:6, size = 1)`, we get:

`"I expected you to assign something to something else with <- where you wrote 123."`

This PR also handles this issue when user submits another function call instead of assigning:

If user code is `sample(1:6, size = 1)` 

And solution code is `a <- sample(1:6, size = 1)`, we get:

`"I expected you to assign something to something else with <- where you called sample()."`


garrettgman (approved):

schloerke (approved):
LGTM given out infix assignment updates that we paired through

PR #156, opened 2020-10-21
correct gradethis:::prep()

### LINKED ISSUE #149: prep() fails when text contains brackets, [ ]
The internal `prep()` function fails when we pass it text that contains brackets:

```r
gradethis:::prep("[aeiou]")
```
```
Error in parse(text = x) : <text>:1:1: unexpected '['
1: [
    ^
```

This prevents us from using `grade_code()` with stringr functions, or other cases where the user is likely to supply a regular expression, which can include `[`.

VincentGuyader:
fix #149  

see <https://github.com/rstudio-education/gradethis/issues/149#issuecomment-713435555>

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -12,9 +12,16 @@ is_pipe <- function(x) {
 .infixes <- c("+", "-", "*", "/", "^", "%%", "%/%", "%in%", .infixes_assign)
 
 is_infix <- function(x, infix_vals = .infixes) {
-  if (is.character(x)) x <- parse(text = x)[[1]]
-  ifelse(is.call(x), as.character(x[[1]]) %in% infix_vals, FALSE)
+  
+  # if str2lang fails, x is not an infix
+  out <- x
+  if (is.character(x)){out<-"";  try(out <- str2lang(x), silent=TRUE) } 
+  # if (is.character(x)) x <- parse(text = x)[[1]]
+  ifelse(is.call(out), as.character(out[[1]]) %in% infix_vals, FALSE)
 }
+
+
+
 
 is_infix_assign <- function(x) {
   is_infix(x, infix_vals = .infixes_assign)
--- a/R/utils.R
+++ b/R/utils.R
@@ -13,11 +13,23 @@ is_pipe <- function(x) {
 
 is_infix <- function(x, infix_vals = .infixes) {
   
-  # if str2lang fails, x is not an infix
-  out <- x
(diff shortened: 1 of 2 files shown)


tests/testthat/test-robust_infix.R
```diff
@@ -0,0 +1,12 @@
+test_that("infix works", {
+  expect_true(is_infix("1+1"))
+  expect_true(is_infix("1*1"))
+  expect_true(is_infix("a %in% b"))
+  expect_false(is_infix("1"))
+
+  })
```
schloerke, line 7:
```suggestion
})

```

tests/testthat/test-robust_infix.R
```diff
@@ -0,0 +1,12 @@
...
+test_that("infix is robust", {
+  expect_false(is_infix("[azeazer]"))
+  expect_false(is_infix("[azeazer"))
+  expect_false(is_infix("[azeazer*"))
+  })
```
schloerke, line 12:
```suggestion
})

```

R/utils.R
```diff
@@ -12,10 +12,17 @@ is_pipe <- function(x) {
...
 
 is_infix <- function(x, infix_vals = .infixes) {
-  if (is.character(x)) x <- parse(text = x)[[1]]
-  ifelse(is.call(x), as.character(x[[1]]) %in% infix_vals, FALSE)
+  
+  # if str2lang fails, x is not an infix
+  out <- x
+  if (is.character(x)){out<-"";  try(out <- str2lang(x), silent=TRUE) } 
+  # if (is.character(x)) x <- parse(text = x)[[1]]
+  ifelse(is.call(out), as.character(out[[1]]) %in% infix_vals, FALSE)
```
schloerke, line 20: Cosmetic changes (untested)

```suggestion
  tryCatch({
    if (is.character(x)) {
      # if str2lang throws, x is not an infix
      out <- str2lang(x)
    } else {
      out <- x
    }
    
    if (!is.call(out)) {
      return(FALSE)
    }
    
    as.character(out[[1]]) %in% infix_vals
  }, error = function(e) {
    # x is not an infix
    FALSE
  })
```

schloerke:
I would not worry about the use of a try/catch in prod. Since we are required to parse unknown input, we need to protect as much as possible.

Great fix!

schloerke (changes requested):
(see suggestions)

schloerke (approved):

Issue #155, opened 2020-10-21
add a parameter to grade_code to disallow partial matching


VincentGuyader:
```
grade_code(allow_partial_matching = FALSE,
  grader_args = list(
    user_quo = expression(read_csv(file = "foo.csv", col_nam = TRUE,quoted_ = TRUE)), 
    solution_quo = expression(read_csv(file = "foo.csv", col_names = TRUE,quoted_na = TRUE))
  )
) 
```

should return :
> This code seems correct, but please write with full parameter(s) names.
> You wrote `col_nam = TRUE` please rewrite with `col_names = TRUE` .
> You wrote `quoted_ = TRUE` please rewrite with `quoted_na = TRUE` .  Give it another try.

VincentGuyader:
I have some issue with unit testing, 

this is ok in interactive mode : 

 ```
 default <- grade_code(
    grader_args = list(
      user_quo = expression(read_csv(file = "foo.csv", col_nam = TRUE,quoted_ = TRUE)), 
      solution_quo = expression(read_csv(file = "foo.csv", col_names = TRUE,quoted_na = TRUE))
    )
  )
  pmatch_FALSE <- grade_code(allow_partial_matching = FALSE,
             grader_args = list(
               user_quo = expression(read_csv(file = "foo.csv", col_nam = TRUE,quoted_ = TRUE)), 
               solution_quo = expression(read_csv(file = "foo.csv", col_names = TRUE,quoted_na = TRUE))
             )
  )
  pmatch_TRUE <- grade_code(allow_partial_matching = TRUE,
             grader_args = list(
               user_quo = expression(read_csv(file = "foo.csv", col_nam = TRUE,quoted_ = TRUE)), 
               solution_quo = expression(read_csv(file = "foo.csv", col_names = TRUE,quoted_na = TRUE))
             )
  )
  expect_true(default$correct)
  expect_true(pmatch_TRUE$correct)
  expect_false(pmatch_FALSE$correct)
  expect_match(object = pmatch_FALSE$message,regexp = "col_names = TRUE")
  expect_match(object = pmatch_FALSE$message,regexp = "quoted_na = TRUE")
```

but not inside `devtools::test()`

the output contains : 

> $correct
> [1] FALSE
> 
> attr(,"class")
> [1] "grader_graded"
> $message
> Your call to read_csv() should include col_names as one of its arguments. You may have misspelled an argument name, or left out an important argument.  Try it again. Perseverence is the key to success.

[...]

PR #160, opened 2020-10-21
fix #148 and road to 155

### LINKED ISSUE #148: Distinguish between written and default args in grade_code feedback
The code below returns a confusing message for students who do not know about the skip argument (an argument that is preset to the default value of 0. `grade_code` adds the argument and the default value ot the student's call before it compares the call to the solution (which also has all missing defaults added to the call).

```r
library(readr)
grade_code(
  grader_args = list(
    user_quo = expression(read_csv("foo.csv")), 
    solution_quo = expression(read_csv("foo.csv", skip = 3))
  )
)
```

````
$message
In read_csv("foo.csv"), I expected 3 where you wrote 0.  Try it again. I have a good feeling about this.
````

We should try to identify which arguments the student conciously wrote and to use that info to construct more transparent feedback messages like

````
$message
In read_csv("foo.csv"), I expected you to add the argument skip = 3.
````

### PR DESCRIPTION
**Pull request by VincentGuyader**
this fix #148 

the main action I made was to add a 'minimalist' parameter ( not sure about the name I choose) to force gradethis:::call_standardise_formals to not  fill out default formals. and I used `minimalist = TRUE` when detecting missing_args 

see : 
 ``` r
library(readr)
user <- rlang::get_expr(quote(read_csv("foo.csv", skip = 3)))
gradethis:::call_standardise_formals(user)
#> read_csv(file = "foo.csv", col_names = TRUE, col_types = NULL, 
#>     locale = default_locale(), na = c("", "NA"), quoted_na = TRUE, 
#>     quote = "\"", comment = "", trim_ws = TRUE, skip = 3, n_max = Inf, 
#>     guess_max = min(1000, n_max), progress = show_progress(), 
#>     skip_empty_rows = TRUE)
gradethis:::call_standardise_formals(user,minimalist = TRUE)
#> read_csv(file = "foo.csv", skip = 3)
```

### DIFF
--- a/R/call_standarise_formals.R
+++ b/R/call_standarise_formals.R
@@ -1,4 +1,4 @@
-call_standardise_formals <- function(code, env = rlang::current_env(),minimalist = FALSE) {
+call_standardise_formals <- function(code, env = rlang::current_env(), include_defaults = TRUE) {
   # try to catch invalid function, i.e., corrupt language object
   tryCatch({
     fxn <- rlang::call_fn(code, env = env)
@@ -11,9 +11,9 @@ call_standardise_formals <- function(code, env = rlang::current_env(),minimalist
   
   
   
-  # standarise, but dont bother trying to fill out default formals
+  # if include_defaults == FALSE standarise, but dont bother trying to fill out default formals
   # - for primitives like mean, unable to distinguish between mean and mean.default
-  if (minimalist | (is_infix(code) || is.primitive(fxn))) {
+  if ( !include_defaults | (is_infix(code) || is.primitive(fxn))) {
     return(rlang::call_standardise(code))
   }
 
--- a/R/call_standarise_formals.R
+++ b/R/call_standarise_formals.R
@@ -13,7 +13,7 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
   
   # if include_defaults == FALSE standarise, but dont bother trying to fill out default formals
   # - for primitives like mean, unable to distinguish between mean and mean.default
-  if ( !include_defaults | (is_infix(code) || is.primitive(fxn))) {
+  if ( (!include_defaults) || is_infix(code) || is.primitive(fxn)) {
     return(rlang::call_standardise(code))
   }
 
--- a/R/call_standarise_formals.R
+++ b/R/call_standarise_formals.R
@@ -1,5 +1,4 @@
-call_standardise_formals <- function(code, env = rlang::current_env()) {
-
+call_standardise_formals <- function(code, env = rlang::current_env(),minimalist = FALSE) {
   # try to catch invalid function, i.e., corrupt language object
   tryCatch({
     fxn <- rlang::call_fn(code, env = env)
@@ -9,9 +8,12 @@ call_standardise_formals <- function(code, env = rlang::current_env()) {
   if (!exists("fxn")) {return(code)} ## some reason the above tryCatch doesn't go to the error part
   if(class(fxn) != "function") {return(code)}
 
+  
+  
+  
   # standarise, but dont bother trying to fill out default formals
-  # for primitives like mean, unable to distinguish between mean and mean.default
-  if (is_infix(code) || is.primitive(fxn)) {
+  # - for primitives like mean, unable to distinguish between mean and mean.default
+  if (minimalist | (is_infix(code) || is.primitive(fxn))) {
     return(rlang::call_standardise(code))
   }
 
--- a/R/detect_mistakes.R
+++ b/R/detect_mistakes.R
@@ -242,8 +242,8 @@ detect_mistakes <- function(user,
   #    will tell them that it expected an na.rm argument, not that na is a surplus
   #    argument.
  
-  explicit_solution <-call_standardise_formals(unpipe_all(answer),env = env,minimalist = TRUE)
-  explicit_user <-call_standardise_formals(unpipe_all(submitted),env = env,minimalist = TRUE)
+  explicit_solution <-call_standardise_formals(unpipe_all(answer),env = env,include_defaults = FALSE)
+  explicit_user <-call_standardise_formals(unpipe_all(submitted),env = env,include_defaults = FALSE)
   explicit_user_names <- real_names(explicit_user)
   explicit_solution_names <-  real_names(explicit_solution)
   missing_args <- explicit_solution_names[!(explicit_solution_names %in% explicit_user_names)]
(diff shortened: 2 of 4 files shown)


VincentGuyader:
as soon as this PR is merged ,it will be straightforward to solve #155 (I hope :) )

garrettgman (changes requested):
@VincentGuyader (and @ColinFay) thank you for tracking this down—I really like the solution. Can we change the `minimalist = FALSE` name to `include_defaults = TRUE`? And adjust `minimalist` to `!include_defaults` accordingly later on? 

I believe this will make the purpose of the argument clear to code readers. I think it would've been easy for me to misunderstand `minimalist` if I didn't have the benefit of your explanation.

VincentGuyader:
done :)

R/call_standarise_formals.R
```diff
@@ -9,9 +8,12 @@ call_standardise_formals <- function(code, env = rlang::current_env()) {
...
 
-  # standarise, but dont bother trying to fill out default formals
-  # for primitives like mean, unable to distinguish between mean and mean.default
-  if (is_infix(code) || is.primitive(fxn)) {
+  
+  
+  
+  # if include_defaults == FALSE standarise, but dont bother trying to fill out default formals
+  # - for primitives like mean, unable to distinguish between mean and mean.default
+  if ( !include_defaults | (is_infix(code) || is.primitive(fxn))) {
```
schloerke, line 16:
```suggestion
  if ( (!include_defaults) || is_infix(code) || is.primitive(fxn)) {
```

R/detect_mistakes.R
```diff
@@ -49,6 +49,7 @@ detect_mistakes <- function(user,
   }
 
   submitted <- user
+  answer <- solution
```
schloerke, line 52: Would prefer to not add unneeded variables.

```suggestion
```

R/detect_mistakes.R
```diff
@@ -236,33 +237,41 @@ detect_mistakes <- function(user,
...
-  user_names <- real_names(user)
-
   # 5. Check that every named argument in the solution appears in the user code.
   #    The outcome of this order is that when a user writes na = TRUE, gradethis
   #    will tell them that it expected an na.rm argument, not that na is a surplus
   #    argument.
-  solution_names <-  real_names(solution) # original solution_names was modified above
-  missing_args <- solution_names[!(solution_names %in% user_names)]
+ 
+  explicit_solution <-call_standardise_formals(unpipe_all(answer),env = env,include_defaults = FALSE)
```
schloerke, line 245:
```suggestion
  explicit_solution <- call_standardise_formals(
    unpipe_all(solution), 
    env = env, 
    include_defaults = FALSE
  )
```

tests/testthat/test_standarize_call_formals.R
```diff
@@ -64,3 +64,25 @@ test_that("When an invalid function passed (i.e., corrupt language object)", {
   testthat::expect_equal(
     call_standardise_formals(user), user)
 })
+
+
+test_that("Standarize call with include_defaults = FALSE", {
+  library(readr)
```
schloerke, line 70: Is there another example that can be tested? 

CRAN will require `readr` to be installed. I'd rather not take a dependency on it.

R/detect_mistakes.R
```diff
@@ -236,33 +237,41 @@ detect_mistakes <- function(user,
...
+  explicit_user <-call_standardise_formals(unpipe_all(submitted),env = env,include_defaults = FALSE)
```
schloerke, line 246:
```suggestion
  explicit_user <- call_standardise_formals(
    unpipe_all(submitted),
    env = env,
    include_defaults = FALSE
  )
```

schloerke (changes requested):

garrettgman (approved):

R/detect_mistakes.R
VincentGuyader: humm, I think we can't do this like that because of line 59 we do :

```
  if (is.call(solution)) {
    solution <- call_standardise_formals(unpipe_all(solution), env = env)
  }
```

so we cant use ` include_defaults = FALSE`  on an already expanded call

R/detect_mistakes.R
VincentGuyader: answer is only needed (at this time) for this instruction : 

```  
explicit_solution <-call_standardise_formals(unpipe_all(answer),env = env,include_defaults = FALSE)
```

I can move up this line to line 52 to avoid variable creation, but it will be less clear than to have this line in step 5 :  

```
  # 5. Check that every named argument in the solution appears in the user code.
  #    The outcome of this order is that when a user writes na = TRUE, gradethis
  #    will tell them that it expected an na.rm argument, not that na is a surplus
  #    argument.
 
  explicit_solution <-call_standardise_formals(unpipe_all(answer),env = env,include_defaults = FALSE)
  explicit_user <-call_standardise_formals(unpipe_all(submitted),env = env,include_defaults = FALSE)
  explicit_user_names <- real_names(explicit_user)
  explicit_solution_names <-  real_names(explicit_solution)
```

tests/testthat/test_standarize_call_formals.R
VincentGuyader: because it's on test/ I think it will only be in Suggest. Not in the Import field of DESCRIPTION file.

I will try to find another example.

R/detect_mistakes.R
schloerke: Ah! Thank you.

Can we rename `answer` to `solution_original`? 🙌

VincentGuyader (commented):
seems ok for me now

R/detect_mistakes.R
```diff
@@ -236,33 +237,45 @@ detect_mistakes <- function(user,
...
-  user_names <- real_names(user)
-
   # 5. Check that every named argument in the solution appears in the user code.
   #    The outcome of this order is that when a user writes na = TRUE, gradethis
   #    will tell them that it expected an na.rm argument, not that na is a surplus
   #    argument.
-  solution_names <-  real_names(solution) # original solution_names was modified above
-  missing_args <- solution_names[!(solution_names %in% user_names)]
+ 
+  explicit_solution <-call_standardise_formals(unpipe_all(solution_original),env = env,include_defaults = FALSE)
```
schloerke, line 245:
```suggestion
  explicit_solution <- call_standardise_formals(
    unpipe_all(solution_original),
    env = env,
    include_defaults = FALSE
  )
```

schloerke (approved):
LGTM pending white space change.

Thank you!

PR #164, opened 2020-11-03
fix #158

### LINKED ISSUE #158: issue with `grade_learnr` unit test
we have an error on check here : 

>-- 1. Failure: Grade learnr check_code (@test_grade_learnr.R#98)  --------------
grade_learnr(user_code = "function(") not equal to grade_feedback(graded(correct = FALSE, message = "The user code of 'function(' produced an parsing error: <text>:2:0: unexpected end of input\n1: function(\n   ^")).
Component "message": 1 string mismatch

> == testthat results  ===========================================================
[ OK: 486 | SKIPPED: 0 | WARNINGS: 0 | FAILED: 1 ]
>1. Failure: Grade learnr check_code (@test_grade_learnr.R#98) 

this is the failing test : 

```
  expect_equal(
    grade_learnr(user_code = "function("),
    grade_feedback(
      graded(
        correct = FALSE,
        message = "The user code of 'function(' produced an parsing error: <text>:2:0: unexpected end of input\n1: function(\n   ^"
      )
    )
  )
```

VincentGuyader:
this fix #158 

I wasn't able to find a commit which was successfull for this test.. 
but it's quite easy to correct it

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -25,7 +25,7 @@ is_infix <- function(x, infix_vals = .infixes) {
       return(FALSE)
     }
     
-    as.character(out[[1]]) %in% infix_vals
+    any(as.character(out[[1]]) %in% infix_vals)
   }, error = function(e) {
     # x is not an infix
     FALSE
--- a/tests/testthat/test_grade_learnr.R
+++ b/tests/testthat/test_grade_learnr.R
@@ -100,7 +100,7 @@ test_that("Grade learnr check_code", {
     grade_feedback(
       graded(
         correct = FALSE,
-        message = "The user code of 'function(' produced an parsing error: <text>:2:0: unexpected end of input\n1: function(\n   ^"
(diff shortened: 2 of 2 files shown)


VincentGuyader:
the issue come from the OS used to launch the test : 

`shQuote("function(")`

return on windows :
image

return on linux 
image

so we need to be less strict on this test

schloerke (changes requested):
Let's switch it to instead change the root cause. `shQuote` is for shell escaping. I think the intent was to use single quotes `sQuote(x, FALSE)`. (Set fancy to FALSE to avoid UTF-8 issues / better reproducibility. 

Thank you!

VincentGuyader:
> 
> 
> Let's switch it to instead change the root cause. `shQuote` is for shell escaping. I think the intent was to use single quotes `sQuote(x, FALSE)`. (Set fancy to FALSE to avoid UTF-8 issues / better reproducibility.
> 
> Thank you!

very nice idea ! (I spend to much time on this!)

i just made the change.

thanks

schloerke (approved):

PR #267, opened 2021-11-13
Add `stage` argument to exercise checker

rossellhayes:
Adds a `stage` argument to `gradethis_exercise_checker()` and `check_exercise()` to take advantage of rstudio/learnr#610.

If `stage` is set, `gradethis` skips parse checking, since parse checking will have already occurred in `learnr`.

### DIFF
--- a/R/gradethis_exercise_checker.R
+++ b/R/gradethis_exercise_checker.R
@@ -155,7 +155,8 @@ check_exercise <- function(
     }
   )
 
-  if (is.null(stage)) {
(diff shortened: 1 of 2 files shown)


gadenbuie (approved):
Maybe we should also check the learnr version and keep the parse checker around for older versions prior to when the learnr parser checker was added?

rossellhayes:
@gadenbuie  That's a good idea. Should the check test for the `learnr` version when the `stage` argument was added, or when the parse checker was added?

PR #289, opened 2022-02-22
Add support for multiple solutions to `code_feedback()`

rossellhayes:
Expands on #286 by adding support for multiple solutions to `code_feedback()`.

When an exercise has multiple solutions, `code_feedback()` now determines the closest solution to provide feedback.

``` r
library(gradethis)

code_feedback("ln(10)", solution_code_all = list("exp(10)", "log(10)"))
#> I expected you to call `log()` where you called `ln()`.

grader <- grade_this({
  pass_if_equal()
  fail(hint = TRUE)
})

grader(
  mock_this_exercise(
    .user_code = "ln(10)",
    .solution_code = 
"# exp ----
exp(10)
# log ----
log(10)"
  )
)
#>   Incorrect. I expected you to call `log()` where you called `ln()`.
#>   Try it again. I have a good feeling about this.
#> >
```

<sup>Created on 2022-02-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/call_standarise_formals.R
```diff
@@ -15,25 +15,73 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
 }
+
+# @staticimports pkg:staticimports
+#   str_extract
+
+call_standardise_keep_partials <- function(code, env = rlang::caller_env()) {
+  tryCatch(
+    rlang::call_standardise(code, env = env),
+    error = function(e) {
+      # Find index of (first) problematic partial match
```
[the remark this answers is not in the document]
gadenbuie: Another edge case, this will also infinitely recurse

```r
call_standardise_keep_partials(quote(runif(1, m = 0, m = 1)))
```

because the error message is

```r
str(e)
#> List of 2
#>  $ message: chr "formal argument \"min\" matched by multiple actual arguments"
#>  $ call   : language match.call(definition, call, expand.dots, envir)
#>  - attr(*, "class")= chr [1:3] "simpleError" "error" "condition"
```

R/call_standarise_formals.R
```diff
@@ -15,25 +15,73 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
+      index <- as.integer(str_extract(e$message, "\\d+")) + 1
+
+      # Recursively run `call_standardise_keep_partials()` on code with
+      # problematic argument removed
+      standardised_call <- call_standardise_keep_partials(code[-index])
```
gadenbuie, line 57:
```suggestion
      # Try to standardize the call while ignoring the un-matchable arg
      standardised_call <- call_standardise_keep_partials(code[-index])
```

R/call_standarise_formals.R
```diff
@@ -15,25 +15,73 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
+
+      # Reassemble original call, by re-adding unaltered problematic arguments
+      # to the standardized call
+      as.call(
+        append(
+          as.list(standardised_call),
+          as.list(code[index]),
+          after = index - 1
+        )
+      )
```
gadenbuie, line 67: I think the reason for this particular formulation (and not `rlang::call_modify()`) is that you're trying to preserve the position of the un-matched arg. (This _why_ would be good to include in the comment, btw.)

But `match.call()` re-arranges the arguments, so if we really want to maintain argument order from the original user formulations we'd have to do a bit more work. I'm on the fence about whether that'd be worth the effort, mostly because while `match.call()` standardizes the call, it doesn't provide a map from original to new arguments.

```r
f <- function(zz = 1, ab = 2, za = 3, ad = 4) NULL

call_standardise_formals(quote(f(ad = 4, z = 1, ab = 2)))
#> f(ab = 2, z = 1, ad = 4)

call_standardise_keep_partials(quote(f(ad = 4, z = 1, ab = 2)))
#> f(ab = 2, z = 1, ad = 4)
```

R/call_standarise_formals.R
```diff
@@ -15,25 +15,73 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
+    }
+  )
+}
+
+call_standardise_formals_recursive <- function(
```
gadenbuie, line 72:
```suggestion
call_standardise_formals_recursive <- function( # nolint
```

R/code_feedback.R
```diff
@@ -129,6 +129,8 @@
 #' @param user_code,solution_code String containing user or solution code. For
 #'   ease of use in `grade_this()`, `.user_code` or `.solution_code` are by
 #'   default retrieved from the calling environment.
+#' @param solution_code_all A list containing the code of all solutions when
```
gadenbuie, line 132: Link to or reference `.solution_code_all`

gadenbuie (commented):
A few initial thoughts, mostly about call standardizing. I haven't looked at the code feedback parts yet

R/call_standarise_formals.R
```diff
@@ -15,25 +15,73 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
 
   ## Add implicit default args to the call
-  rlang::call_standardise(
+  call_standardise_keep_partials(
     rlang::call_modify(code_std, !!!args_default[args_default_missing]),
     env = env
   )
 }
+
+# @staticimports pkg:staticimports
```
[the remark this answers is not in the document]
rossellhayes: Moved in [1c2d7fa](https://github.com/rstudio/gradethis/pull/289/commits/1c2d7facc2dd0715c2dac6416c6c3afb21131884)

R/code_feedback.R
rossellhayes: Documented in [8c143b7](https://github.com/rstudio/gradethis/pull/289/commits/8c143b7b9aa0ebc66deb8059b48f02438244e62a)

R/call_standarise_formals.R
rossellhayes: So the real reason this was using `append()` instead of `call_modify()` is that I didn't know `call_modify()` existed. Switch to `rlang::call_modify()` in [496b533](https://github.com/rstudio/gradethis/pull/289/commits/496b53345e7bb98d0952c3f365cb1ba4b4cd2816).

It would be ideal to keep the order that the student gave their arguments, but unless you can think of a simple solution, that seems like a lot more work for a small benefit.

R/call_standarise_formals.R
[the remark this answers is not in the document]
rossellhayes: I _believe_ both these cases are covered by changes in [1731114](https://github.com/rstudio/gradethis/pull/289/commits/1731114069e27376f574be2267ea408be3731587)

R/graded.R
```diff
@@ -650,15 +650,15 @@ fail_if_code_feedback <- function(
...
-      is_placeholder(solution_code) ||
-        is_missing(solution_code) ||
-        is.null(solution_code) ||
-        length(solution_code) == 0 ||
-        !nzchar(solution_code)
+      is_placeholder(solution_code_all) ||
+      is_missing(solution_code_all) ||
+      is.null(solution_code_all) ||
+      length(solution_code_all) == 0 ||
+      !nzchar(solution_code_all)
```
gadenbuie: This needs a little more refactoring. `solution_code_all` is probably a list with class `"gradethis_solutions"`. If it's a character, we'd want to turn it into a `gradethis_solutions` (maybe?). But the check needs to be done a little differently.

Also, since we might end up calling `solutions_prepare()` on an already prepared `solution_code_all`, we should add an escape hatch there to return the object directly if it inherits the `gradethis_solutions` class.

tests/testthat/helper-expect.R
```diff
@@ -173,12 +194,19 @@ expect_feedback <- function(
...
 
 
-create_learnr_env <- function(user_code, solution_code = NULL, envir_prep, eval = TRUE) {
+create_learnr_env <- function(
+  user_code,
+  solution_code = NULL,
+  solution_code_all = NULL,
+  envir_prep,
+  eval = TRUE
+) {
```
gadenbuie, line 203: Not that we need to do it now, but it'd be nice to refactor these tests now that we have `mock_this_exercise()` and `learnr:::mock_exercise()`.

tests/testthat/test-code_feedback.R
```diff
@@ -14,6 +14,19 @@ test_that("code_feedback() returns a string if there are differences or NULL", {
...
+  expect_equal(
+    code_feedback("a", gradethis_solutions("aa", "bb")),
+    "I expected `aa` where you wrote `a`."
+  )
+
+  # If there's a tie, the first option is selected
+  expect_equal(
+    code_feedback("a", gradethis_solutions("b", "c")),
+    "I expected `b` where you wrote `a`."
+  )
```
gadenbuie, line 27: See previous comments. Thinking about this again, we have three cases:

1. Tie for all solutions — use last
2. Tie among some solutions — use first? unless the last one is included?
3. Clear winner (by any amount) — use winner.

The middle case is the trickiest.

R/code_feedback.R
```diff
@@ -218,6 +235,47 @@ with_maybe_code_feedback <- function(val, expr) {
...
+      paste(collapse = "\n")
+  }
+
+  user_code <- standardise_code_text(user_code)
+
+  solution_code_all <- solution_code_all %>%
+    purrr::map_chr(standardise_code_text)
+
+  # Find the index of the solution code that the user code is closest to
+  which.min(utils::adist(user_code, solution_code_all))
```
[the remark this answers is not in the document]
rossellhayes: If there is a tie, `which.min()` returns the first index in the tie, e.g.
``` r
options <- c("abc", "y", "z")
which.min(adist("x", options))
#> [1] 2
```

I don't know about returning `integer(0)` if there's a tie, because we would lose information if e.g. there were five options and the tie was only between two of them.

Since `.solution` defaults to the last value in `.solution_code_all()`, we could have it return the last tie index instead of the first tie index. That way if `.solution` is one of the tied values, that would always be the one returned. If `.solution` isn't one of the tied values, it seems reasonable to choose one arbitrarily.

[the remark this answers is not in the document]
gadenbuie: Sorry, I realized later ([last comment](https://github.com/rstudio/gradethis/pull/289#discussion_r818142324)) that there are two tie scenarios. We'd only return `integer(0)` if the tie is between _all options_.

Otherwise, we'd return the first in the tie — unless the tie includes `.solution` (i.e. the last value).

So yeah, this would all require replacing `which.min()` with our own logic.

R/code_feedback.R
```diff
@@ -218,6 +235,47 @@ with_maybe_code_feedback <- function(val, expr) {
   )
 }
 
+which_closest_solution_code <- function(
+  user_code = .user_code,
+  solution_code_all = .solution_code_all
+) {
```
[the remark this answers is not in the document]
rossellhayes: Removed in [1248317](https://github.com/rstudio/gradethis/pull/289/commits/1248317a8ebcebc142438e4378c0a6c2c51e2b05)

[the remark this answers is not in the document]
gadenbuie: sorry I should have clarified that we should also remove the assertions and checks around the placeholders, too

R/code_feedback.R
```diff
@@ -175,11 +179,24 @@ code_feedback <- function(
...
+        assert_not_placeholder(solution_code)
+      }
+
+      solution_code_all <- solutions_prepare(solution_code)
+    }
   }
 
+  # Figure out which solution code is .solution_code here
+  closest_solution <- which_closest_solution_code(user_code, solution_code_all)
+  solution_code <- solution_code_all[[closest_solution]]
```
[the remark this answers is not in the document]
rossellhayes: I think this is handled by [1d86cf0](https://github.com/rstudio/gradethis/pull/289/commits/1d86cf03a8edba9e2c69bf9531507e6aa2f8ecd0).

If `which_closest_solution()` gives an invalid index, it now falls back to the last index in `solution_code_all`, which should be equivalent to `solution_code`.

R/code_feedback.R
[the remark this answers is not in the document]
rossellhayes: Is it okay to always return the _last_ value if there's a tie? If we can do that, we could simplify the logic, because `.solution` is always the last value.

R/code_feedback.R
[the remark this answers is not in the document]
rossellhayes: I should've caught that 🤦‍♂️
[d4f0488](https://github.com/rstudio/gradethis/pull/289/commits/d4f04889899d0ece2a66051c682e74b6a0ff36f8)

R/code_feedback.R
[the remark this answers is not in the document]
gadenbuie: > Is it okay to always return the _last_ value if there's a tie?

No, just when the last value is included in the tie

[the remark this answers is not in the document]
rossellhayes: Implemented in [658ef2c](https://github.com/rstudio/gradethis/pull/289/commits/658ef2c58afd757863580ea0cc937e8c8f0d4371), but we can discuss further if needed

tests/testthat/test-code_feedback.R
rossellhayes: [658ef2c](https://github.com/rstudio/gradethis/pull/289/commits/658ef2c58afd757863580ea0cc937e8c8f0d4371) implements this solution:
1. Tie for all solutions — use last
2. Tie among some solutions — use last
3. Clear winner (by any amount) — use winner

R/graded.R
rossellhayes: [d1f0ed3](https://github.com/rstudio/gradethis/pull/289/commits/d1f0ed3b591e115c56a1f5e861e7b6b09385dd03) adds an early return to `solutions_prepare()`

R/call_standarise_formals.R
```diff
@@ -15,25 +15,135 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
+  # If the function from the user code cannot be found, early exit because
+  #   standardization is impossible
+  fn <- tryCatch(call_fn(code, env), error = as.null)
+  if (is.null(fn)) return(code)
+
+  tryCatch(
+    rlang::call_match(code, fn),
+    error = function(e) {
+      # Check that error is caused by an ambiguous partial argument
+      # If not, return the code unaltered
```
gadenbuie, line 55:
```suggestion
      # Check that error is caused by an ambiguous partial argument, which we
      # are forced to identify from its error message. The \Q...\E regex meta
      # characters mark the text inside them as literal. This way we can use the
      # literal, translated error message, but replace the `%d` with `\d+`.
```

R/call_standarise_formals.R
```diff
@@ -15,25 +15,135 @@ call_standardise_formals <- function(code, env = rlang::current_env(), include_d
...
+          pattern = "%d",
+          replacement = "\\E\\d+\\Q",
+          x = gettext("argument %d matches multiple formal arguments", domain = "R"),
+          fixed = TRUE
+        ),
+        "\\E"
+      )
+      if (!grepl(pattern, e$message)) {
+        return(code)
+      }
```
gadenbuie, line 68: Will this end up swallowing other kinds of errors?

R/code_feedback.R
```diff
@@ -218,6 +234,52 @@ with_maybe_code_feedback <- function(val, expr) {
...
+    closest_solution <- length(solution_code_all)
+  }
+
+  closest_solution
+}
+
+which.min.last <- function(x) {
+  x <- rev(x)
+  index <- which.min(x)
+  rev(seq_along(x))[index]
```
gadenbuie, line 280: Is there an advantage to using `which.min()`? This is more verbose but less mind-bendy with the reversing:

```r
idx_min <- which(x == min(x))

if (length(idx_min) < 2) {
  return(idx_min)
}

if (length(x) %in% idx_min) {
  # return last solution when it's in the tie
  return(length(x))
}

idx_min[1]
```

Also the version above returns the _final_ solution if it's in the tie, not the last solution _in the tie_. Basically: we take the final solution whenever it's in the possibilities because it's the default solution. But if it's not the final solution we'd use the first one on the list.

gadenbuie: That's different enough from `which.min.last()` that maybe we'd want to move it back into `which_closest_solution_code()`.

R/graded.R
```diff
@@ -650,15 +650,15 @@ fail_if_code_feedback <- function(
...
-      is_placeholder(solution_code) ||
-        is_missing(solution_code) ||
-        is.null(solution_code) ||
-        length(solution_code) == 0 ||
-        !nzchar(solution_code)
+      is_placeholder(solution_code_all) ||
+        is_missing(solution_code_all) ||
+        is.null(solution_code_all) ||
+        length(solution_code_all) == 0 ||
+        !nzchar(solution_code_all)
```
gadenbuie, line 661:
```suggestion
        any(!nzchar(solution_code_all))
```

Surprisingly `nzchar()` works on a list

```r
!nzchar(list('a', 'b', ''))
#> [1] FALSE FALSE  TRUE
```

but we need to make sure at least one solution has some code in it. (That seems unlikely but it is certainly the safest thing for this test.)

R/call_standarise_formals.R
gadenbuie: TIL about `\Q...\E` ! That's a pretty neat trick, but it took me a bit to work it out.

R/code_feedback.R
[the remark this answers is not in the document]
rossellhayes: @gadenbuie To clarify, if we have a vector with length 5 and the tie is between 2 and 4, is it necessary to return 2, or is it also acceptable to return 4? If it is okay to return 4, then it would simplify the logic, because a tie including between 2 and 5 (the last value) wouldn't need to be treated any different than a tie between 2 and 4.

R/call_standarise_formals.R
rossellhayes: Yes, the idea is that if we have any kind of parsing error besides a "argument x matched multiple formal arguments" error, it will just treat the code as unstandardizable and return it as is. Are there any other errors that we'd like to try to recover from?

R/code_feedback.R
rossellhayes: This seems like a good solution. The advantage of `which.min()` is that it's very fast, but exercise evaluation is a pretty slow process anyway, so I don't think shaving off a few milliseconds is worth less readable code.

R/code_feedback.R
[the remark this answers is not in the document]
gadenbuie: Yeah I thought we would return 2. It's a little convoluted, but I think ranking solutions top-bottom (top first) makes sense — other than the special case of the "default" solution.

I'd rather make the first solution the default than generally go with a bottom-top preference order.

R/call_standarise_formals.R
gadenbuie: That makes sense, I just wanted to make sure we've accounted for upstream places that might want the standardized code but I think they're caught by other mistake detectors... here's how I convinced myself.

```r
dummy <- function(aaa = 1, aaaa = 2) {
  aa + aaa * 10
}

grade_this_code()(mock_this_exercise("dummy(a = 1, aa = 2)", "dummy(2, 1)"))
#> <gradethis_graded_this_code: [Incorrect]
#>   `dummy()` accepts an argument named `"aaa"`. More than one of your
#>   argument names in `dummy()` will be matched to `"aaa"`, which will
#>   cause an error. Try writing out the full argument names. Try it
#>   again. Perseverence is the key to success.
#> >

call_standardise_keep_partials(quote(dummy(a = 1, aa = 2)))
#> dummy(a = 1, aa = 2)
```

R/code_feedback.R
rossellhayes: I implemented this in [5c11542](https://github.com/rstudio/gradethis/pull/289/commits/5c1154253e871364e291edbf1362519c18fc8f32). The result is now:
1. Tie for all solutions – use last
2. Tie among some solutions, including the last solution – use last
3. Tie among some solutions, not including the last solution – use earliest solution included in tie
4. Clear winner (by any amount) – use winner

gadenbuie (approved):
This is excellent, thank you! I think the implementation is ready to go!

We could use more documentation around multiple solutions. I don't mind if you do that in a separate PR since this one has quite a lot of comments and commits. We should work on it soon, though, since we're likely to uncover small changes in that process.

rossellhayes:
The failing check is a [known issue in `rmarkdown`](https://github.com/rstudio/rmarkdown/issues/2324), so I'm going to merge.

PR #296, opened 2022-03-10
Add support for multiple solutions with different results

### LINKED ISSUE #293: Add support for multiple solutions in `pass_if_equal()`
Currently, multiple solutions (as implemented in #289) only supports multiple solution _code_. We need to implement multiple solution results, which would be accessible from functions like `pass_if_equal()`.

rossellhayes:
- Adds `.solution_all` to check env, a `gradethis_solutions` list of the results of `.solution_code_all`
	- `.solution_all` is created with `delayedAssign()`, so results are only evaluated as needed
- Updated `pass_if_equal()` to support multiple solutions
	- If `y` is a `gradethis_solutions` object (e.g. `solution_all`), `pass_if_equal()` passes if `x` matches _any_ of the values stored in `y`
	- By default, `pass_if_equal()` still uses `y = .solution`, _not_ `y = .solution_all`. This is to prevent unnecessary evaluation if all elements of `.solution_code_all` have the same result. You should only use `y = .solution_all` if the elements of `.solution_code_all` have _different_ results.

``` r
library(gradethis)

.solution_code <- "# exp ----
exp(10)
# log ----
log(10)"

ex <- mock_this_exercise(.user_code = "ln(10)", !!.solution_code)
ex$.solution_all
#> # log --------------------------------------------------------------------------
#> 
#> 2.30258509299405
#> 
#> # exp --------------------------------------------------------------------------
#> 
#> 22026.4657948067

# User code doesn't match
grade_this({pass_if_equal(.solution_all); fail()})(
  mock_this_exercise(.user_code = "ln(10)", !!.solution_code)
)
#>   Incorrect. I expected you to call `log()` where you called `ln()`.
#>   Don't give up now, try it one more time.
#> >

# User code matches first solution
grade_this({pass_if_equal(.solution_all); fail()})(
  mock_this_exercise(.user_code = "exp(10)", !!.solution_code)
)
#> <gradethis_graded: [Correct] Well done! Correct!>

# User code matches second solution
grade_this({pass_if_equal(.solution_all); fail()})(
  mock_this_exercise(.user_code = "log(10)", !!.solution_code)
)
#> <gradethis_graded: [Correct] Swell job! Correct!>
```

<sup>Created on 2022-03-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Closes #293 and #299.


gadenbuie (approved):
Looks great, I think this is a great start for supporting multiple solutions!

I made a couple edits and a few small changes; if those look good to you then this is ready to merge!

rossellhayes:
@gadenbuie This looks great! `solution_code_closest()` is a nice addition, and I hadn't seen `purrr::partial()` before, but that's a great tool to know about

Issue #345, opened 2023-05-04
`pass_if_equal()` does not pass when `.result` and `.solution` are the same `ggplot`


rossellhayes:
`pass_if_equal()` detects a difference between two identical `ggplot`s if one has been printed and one has not.
Because `.result` is shown to the student and `.solution` is not, this will almost always be the case.

``` r
library(ggplot2)
library(gradethis)

plot_code <- quote(
  ggplot(mpg, aes(displ, hwy, colour = class)) + 
    geom_point()
)

.result   <- eval(plot_code)
.solution <- eval(plot_code)

# Passes if both plots have not been print
pass_if_equal()
#> <gradethis_graded: [Correct] That's a first-class answer! Correct!>

.result
```

![](https://i.imgur.com/IhgCZ1e.png)

``` r

# Fails if one plot has been printed
pass_if_equal()
#> NULL

.solution
```

![](https://i.imgur.com/np11qpU.png)

``` r

# Passes if both plots have been printed
pass_if_equal()
#> <gradethis_graded: [Correct] Superb work! Correct!>
```

<sup>Created on 2023-05-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This is because printing a `ggplot` adds additional computed data to each layer:

``` r
library(ggplot2)
library(gradethis)

plot_code <- quote(
  ggplot(mpg, aes(displ, hwy, colour = class)) + 
    geom_point()
)

.result   <- eval(plot_code)
.solution <- eval(plot_code)

.result
```

![](https://i.imgur.com/6dbX1kb.png)

``` r

waldo::compare(.result, .solution)
#> `old$layers[[1]]` is length 15
#> `new$layers[[1]]` is length 12
#> 
#>     names(old$layers[[1]]) | names(new$layers[[1]])    
#> [1] "aes_params"           | "aes_params"           [1]
#> [2] "computed_geom_params" -                           
#> [3] "computed_mapping"     -                           
#> [4] "computed_stat_params" -                           
#> [5] "constructor"          | "constructor"          [2]
#> [6] "data"                 | "data"                 [3]
#> [7] "geom"                 | "geom"                 [4]
#> 
#> `old$layers[[1]]$computed_geom_params` is a list
#> `new$layers[[1]]$computed_geom_params` is absent
#> 
#> `old$layers[[1]]$computed_mapping` is an S3 object of class <uneval>, a list
#> `new$layers[[1]]$computed_mapping` is absent
#> 
[...]

PR #346, opened 2023-05-04
Add `fail_if_not_equal()` and `gradethis_equal()`

rossellhayes:
- Adds `fail_if_not_equal()`.
- `pass_if_equal()`, `fail_if_equal()`, and `fail_if_not_equal()` now call `gradethis_equal()`, an S3 generic which calls `waldo::compare()` by default. This allows other methods for `gradethis_equal()` to handle special cases, like #345.


nischalshrestha (approved):
Nice work! This looks like it will be quite useful for custom comparison of objects if waldo is not sufficient.

PR #347, opened 2023-05-12
Add default arguments to `gradethis_equal()`

rossellhayes:
- `gradethis_equal()` now has default values of `x = .result` and `y = .solution`.
- `gradethis_equal.default()` now has a default value of `tolerance = sqrt(.Machine$double.eps)`.
	- This fixes a bug where `gradethis_equal()` would fail if called without explicitly setting a `tolerance` argument.


nischalshrestha (approved):
Good catch!

PR #353, opened 2023-05-17
Ignore differences between `ggplot2` arguments like `color` vs. `colour` in `code_feedback()`

### LINKED ISSUE #342: Add special handling for `color` vs. `colour` in automated code grading
If a function accepts both `color` and `colour` as arguments (e.g. ggplot2 functions), `detect_mistakes()` should include some special handling so student's don't get feedback like "I expected you to use `color = "blue"` where you had `colour = "blue"`."

rossellhayes:
`call_standardise_formals()` now applies `ggplot2::standardise_aes_names()` to all arguments of `ggplot2` functions. This means it no longer detects inconsequential differences between e.g. `color =` vs `colour =`

``` r
library(gradethis)
library(ggplot2)

code_feedback(
  "ggplot(mpg, aes(displ, hwy, color = class)) + geom_point()",
  "ggplot(mpg, aes(displ, hwy, colour = class)) + geom_point()"
)
#> NULL
```

<sup>Created on 2023-05-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Closes #342.


nischalshrestha (approved):
Nice!

